PR #11, opened 2015-06-25
Rewrite rules modified to convert(to and from) algebricks function

shivanimall:


prestoncarman:
Oops, see my comments on the old pull request.

PR #12, opened 2015-06-26
Node-Before-After files to be added to the codebase

shivanimall:


vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-operators.xml
```diff
@@ -263,10 +263,10 @@
 
     <!-- op:is-same-node($parameter1 as node(), $parameter2 as node()) as xs:boolean -->
     <operator name="op:is-same-node">
-        <param name="parameter1" type="node()?"/>
-        <param name="parameter2" type="node()?"/>
+        <param name="parameter1" type="node()"/>
+        <param name="parameter2" type="node()"/>
         <return type="xs:boolean"/>
-    </operator>
+     </operator>
```
prestoncarman: Why the extra space?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/node/AbstractNodePositionalCheckEvaluator.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private final NodeTreePointable ntp1 = (NodeTreePointable) NodeTreePointable.FACTORY.createPointable();
+    private final NodeTreePointable ntp2 = (NodeTreePointable) NodeTreePointable.FACTORY.createPointable();
+    private final TypedPointables tp = new TypedPointables();
+
+    @Override
+    protected void evaluate(TaggedValuePointable[] args, IPointable result) throws SystemException {
+
+        parameterTypeCheck(args[0], ntp1);
```
prestoncarman: I think it makes the code cleaner and short if you don't have the function call and just have two if statements like the following:

if (args[0].getTag() != ValueTag.NODE_TREE_TAG) {
    throw new SystemException(ErrorCode.FORG0006);
}

prestoncarman:
Only two requests.

shivanimall:
My bad. Previously made changes did not apply on this branch. Code updated. Thanks!

PR #20, opened 2015-07-14
fn:data added during translation

shivanimall:


vxquery-core/src/main/java/org/apache/vxquery/xmlquery/translator/XMLQueryTranslator.java
```diff
@@ -1969,6 +1970,13 @@ private ILogicalExpression normalize(ILogicalExpression expr, SequenceType type)
         }
     }
 
+    private ILogicalExpression data(ILogicalExpression expr) throws SystemException {
+        ILogicalExpression atomizedExpr = new ScalarFunctionCallExpression(BuiltinFunctions.FN_DATA_1,
```
prestoncarman: Since this function is one statement, remove the variable and make that the return statement.

prestoncarman:
Two short comments.

prestoncarman:
I did not pull the count change into master. Please include that commit in a later patch so it can be reviewed.

PR #23, opened 2015-08-10
Prestonc/xmark fixes

prestoncarman:
Includes a bunch of little fixes for XMark queries: 
- command line comments are more standardized.
- Added algebricks rules to support more complex unnesting.
- Moved customized algebricks rules that a special folder.
- Updated algebricks conversion rule to ignore general comparisons.
- Added used variable check a rule.
- Fixed general comparison for nodes.
- IF statement now only evaluates the true result.
- Updated JUnit tests.


westmann:
I realize that this is not directly related to these changes, but I'll ask anyway.
Is there a good way that we can add tests for these specific changes that
a) would guide a reviewer to understand the change in behavior and
b) protect us from regressions that change the plans but not the result?

prestoncarman:
Most of these changes are from slowly fixing each issue that came up while investigating XMark q08. Each fix lead to a new error that lead to a new fix. The XQTS and VXQuery test suite were used to make sure the change did not adversely affect the whole system. In fact, XQTS now has ~150 more passing queries. Since our current testing model only uses query and results, its possible to test some of these concepts independently but not all. 
Currently, VXQuery does not have a query plan test suite. Although this could help test some of the changes to the system, I think we still have a long way to go to get a simplified optimized query plan. Query plan checking may require a lot of changes as we continue to expand support for XQuery. It would require thought as to what query plans we want to test and will not have future changes to their result (although this may not be avoidable).
As we start to get XMark queries working, they could be included into the VXQuery Test Suite to help make sure future changes do not break them.

vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/algebricksalternatives/ExtractFunctionsFromJoinConditionRule.java
```diff
@@ -0,0 +1,141 @@
...
+import edu.uci.ics.hyracks.algebricks.core.algebra.functions.FunctionIdentifier;
+import edu.uci.ics.hyracks.algebricks.core.algebra.operators.logical.AbstractBinaryJoinOperator;
+import edu.uci.ics.hyracks.algebricks.core.algebra.operators.logical.AbstractLogicalOperator;
+import edu.uci.ics.hyracks.algebricks.core.algebra.operators.logical.AssignOperator;
+import edu.uci.ics.hyracks.algebricks.core.algebra.operators.logical.visitors.VariableUtilities;
+import edu.uci.ics.hyracks.algebricks.core.rewriter.base.IAlgebraicRewriteRule;
+
+/**
+ * Factors out function expressions from each comparison function or similarity function in join condition by assigning them to a variables, and replacing the function expressions with references to those variables.
+ * Examples:
```
sjaco002, line 44: Is this file formatted? This comment seems to be really long

vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/algebricksalternatives/ExtractFunctionsFromJoinConditionRule.java
```diff
@@ -0,0 +1,141 @@
...
+                        context.computeAndSetTypeEnvironmentForOperator(newAssign);
+                        context.computeAndSetTypeEnvironmentForOperator(joinOp);
+                    }
+                }
+            }
+            return modified;
+        } else {
+            return false;
+        }
+    }
```
sjaco002, line 139: Why do you want to pull the functions out as assigns? Is this to provide an optimizable format for later rules?

vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/algebricksalternatives/MoveFreeVariableOperatorOutOfSubplanRule.java
```diff
@@ -0,0 +1,178 @@
...
+                //Make op2 the input of the subplan
+                Mutable<ILogicalOperator> op2OutsideRef = op0.getInputs().get(0);
+                op2OutsideRef.setValue(op2);
+
+                return true;
+            }
+
+        }
+        return false;
+    }
```
sjaco002, line 176: We have a very similar rule to this in Hyracks (Movefreevariableoperatoroutofsubplanrule). Is there a reason to have these rules separated?

sjaco002:
Just a few comments.

vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/algebricksalternatives/MoveFreeVariableOperatorOutOfSubplanRule.java
prestoncarman, line 176: I add support for moving subplan out. There is a pending patch to Hyracks to include this functionality. In the future we will be able to remove this rule from VXQuery.

vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/algebricksalternatives/ExtractFunctionsFromJoinConditionRule.java
prestoncarman, line 139: Yes, this helps with other optimization rules that require only variables in the condition.

vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/algebricksalternatives/ExtractFunctionsFromJoinConditionRule.java
prestoncarman, line 44: There is a pending patch to Hyracks to include this functionality. In the future we will be able to remove this rule from VXQuery. Also the patch fixes the comment spacing.

prestoncarman:
The changes have been merged with master.

PR #31, opened 2016-03-03
VXQUERY-184: fixed the order by translator

pavlopoul:
We updated the order by clause to use the atomic values for sorting


prestoncarman:
Nice simple change.

prestoncarman:
Merged with master.

PR #33, opened 2016-03-11
Solved: Issue with "compileonly" flag while executing queries

ankitladhania:
Issue No. : VXQUERY-139
Description : Previously executing queries with compileonly flag
gave NPE, which was due to null value of IHyracksClientConnection.

Solver : @ankitladhania

### DIFF
--- a/vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
+++ b/vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
@@ -248,7 +248,7 @@ public class VXQuery {
             }
             return nodeList;
         }
-        return new String[0];
+        return null;
     }
 
     /**
--- a/vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
+++ b/vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
@@ -240,13 +240,16 @@ public class VXQuery {
      * @throws Exception
      */
     private String[] getNodeList() throws Exception {
-        Map<String, NodeControllerInfo> nodeControllerInfos = hcc.getNodeControllerInfos();
-        String[] nodeList = new String[nodeControllerInfos.size()];
-        int index = 0;
-        for (String node : nodeControllerInfos.keySet()) {
-            nodeList[index++] = node;
+        if(hcc != null) {
+            Map<String, NodeControllerInfo> nodeControllerInfos = hcc.getNodeControllerInfos();
+            String[] nodeList = new String[nodeControllerInfos.size()];
+            int index = 0;
+            for (String node : nodeControllerInfos.keySet()) {
+                nodeList[index++] = node;
+            }
+            return nodeList;
         }
-        return nodeList;
+        return new String[0];
     }
 
     /**
--- a/vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
+++ b/vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
@@ -81,8 +81,7 @@ public class VXQuery {
     /**
      * Constructor to use command line options passed.
      *
-     * @param opts
-     *            Command line options object
+     * @param opts Command line options object
      */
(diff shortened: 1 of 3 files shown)


vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
```diff
@@ -240,13 +240,16 @@ private void runQueries() throws IOException, SystemException, Exception {
      * @throws Exception
      */
     private String[] getNodeList() throws Exception {
-        Map<String, NodeControllerInfo> nodeControllerInfos = hcc.getNodeControllerInfos();
-        String[] nodeList = new String[nodeControllerInfos.size()];
-        int index = 0;
-        for (String node : nodeControllerInfos.keySet()) {
-            nodeList[index++] = node;
+        if(hcc != null) {
```
prestoncarman: Please format the code according to our standard. You can load the eclipse format file into your IDE.

vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
```diff
@@ -240,13 +240,16 @@ private void runQueries() throws IOException, SystemException, Exception {
...
+            Map<String, NodeControllerInfo> nodeControllerInfos = hcc.getNodeControllerInfos();
+            String[] nodeList = new String[nodeControllerInfos.size()];
+            int index = 0;
+            for (String node : nodeControllerInfos.keySet()) {
+                nodeList[index++] = node;
+            }
+            return nodeList;
         }
-        return nodeList;
+        return new String[0];
```
prestoncarman: Why use an empty array instead of null?

prestoncarman:
Thanks for submitting a fix. Take a look at [1] to set up your IDE to format the code.

[1] http://vxquery.apache.org/development_eclipse_setup.html

vxquery-cli/src/main/java/org/apache/vxquery/cli/VXQuery.java
ankitladhania: When I was using null the code was giving NPE as below : 

`Exception in thread "main" java.lang.NullPointerException 
at org.apache.vxquery.metadata.VXQueryMetadataProvider.getClusterLocations(VXQueryMetadataProvider.java:102)
    at org.apache.vxquery.metadata.VXQueryMetadataProvider.getClusterLocations(VXQueryMetadataProvider.java:97)
    at org.apache.vxquery.xmlquery.query.XMLQueryCompiler.<init>(XMLQueryCompiler.java:182)
    at org.apache.vxquery.cli.VXQuery.runQueries(VXQuery.java:264)
    at org.apache.vxquery.cli.VXQuery.execute(VXQuery.java:161)
    at org.apache.vxquery.cli.VXQuery.main(VXQuery.java:121)`

prestoncarman: The change may require other places to be updated to support a null value. I was not sure how much would need to be updated.

ankitladhania: Should I update all the places that require change ?

prestoncarman: Yes, if all these changes are only in VXQuery (I think this is the case, but not sure). If Hyracks requires a empty array, then use the array.

ankitladhania:
Hi Preston,
Can you please tell me about the command for test suite of the project ? 
The docs are incorrect.

Regards,
Ankit

ankitladhania:
@ecarm002 Please look at the above PR

ankitladhania:
Hi Preston,
I have made changes as discussed please review it.

Regards,
Ankit

prestoncarman:
Thanks for fixing the formatting. I liked the first two commits, so I squashed them and merged them into master. After looking at the third in more detail, I found it was extra code that did not improve the code base. Please confirm it's working and comment on the issue.

PR #36, opened 2016-05-16
The object datamodel implementation: ObjectBuilder and ObjectPointable

riyafa:


vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,128 @@
...
+    private static int getSlotValue(byte[] bytes, int start, int idx) {
+        return IntegerPointable.getInteger(bytes, getSlotArrayOffset(start) + idx * SLOT_SIZE);
+    }
+
+    private static int getEntryCount(byte[] bytes, int start) {
+        return IntegerPointable.getInteger(bytes, start);
+    }
+
+    private static int getKeyLength(byte[] b, int s) {
+        return ((b[s + 1] & 0xff) << 8) + ((b[s + 2] & 0xff) << 0) + 3;
```
prestoncarman, line 64: The key is defined by the UTF8StringPointable. Use the pointable's methods to determine the key's length. This way in the future, if the UTF8StringPointable changes the way length is defined, you do not need to change this code.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.hyracks.data.std.primitive.UTF8StringPointable;
+import org.apache.hyracks.data.std.primitive.VoidPointable;
+import org.apache.hyracks.data.std.util.ArrayBackedValueStorage;
+import org.apache.vxquery.datamodel.accessors.TaggedValuePointable;
+import org.apache.vxquery.datamodel.builders.sequence.SequenceBuilder;
+import org.apache.vxquery.exceptions.ErrorCode;
+import org.apache.vxquery.exceptions.SystemException;
+import org.apache.vxquery.runtime.functions.util.FunctionHelper;
+
+public class ObjectPointable extends AbstractPointable {
```
prestoncarman, line 35: Add a comment that explains the data model. Also, do you want to have the string tag with the key?

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,128 @@
...
+        int dataAreaOffset = getDataAreaOffset(bytes, start);
+        int entryCount = getEntryCount();
+        int s, l;
+        for (int i = 0; i < entryCount; i++) {
+            s = dataAreaOffset + getRelativeEntryStartOffset(i);
+            l = getKeyLength(bytes, s);
+            key1.set(bytes, s, l);
+            if (comparePointable(key1, key)) {
+                int m = getEntryLength(i);
+                pointer.set(bytes, s + l, getEntryLength(i) - l);
```
prestoncarman, line 108: Are you returning the tagged value?

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,128 @@
...
+                break;
+            } else {
+                //Todo: throw an exception
+            }
```
prestoncarman, line 112: Do you need this else?

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,128 @@
...
+    private int getRelativeEntryStartOffset(int idx) {
+        return idx == 0 ? 0 : getSlotValue(bytes, start, idx - 1);
+    }
+
+    private int getEntryLength(int idx) {
+        return getSlotValue(bytes, start, idx) - getRelativeEntryStartOffset(idx);
+    }
+
+    private boolean comparePointable(IPointable p1, IPointable p2) {
+        return FunctionHelper.arraysEqual(p1.getByteArray(), p1.getStartOffset(), p1.getLength(), p2.getByteArray(),
```
prestoncarman, line 125: I added a new helper function that compares pointables. Use it.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/ValueTag.java
```diff
@@ -82,6 +82,7 @@
     public static final int COMMENT_NODE_TAG = 105;
     public static final int PI_NODE_TAG = 106;
     public static final int NODE_TREE_TAG = 107;
+    public static final int OBJECT_TAG=109;
```
prestoncarman, line 85: Should this be JSON_OBJECT_TAG? I am wondering if its better to add JSON or leave it off?

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/XDMConstants.java
```diff
@@ -32,6 +32,8 @@
 
     private static final byte[] EMPTY_SEQUENCE;
 
+    private static final byte[] EMPTY_OBJECT;
```
prestoncarman, line 35: Do you think we need this constant? How often will you have an empty object as a result?

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/XDMConstants.java
```diff
@@ -19,11 +19,11 @@
...
-import org.apache.vxquery.datamodel.builders.sequence.SequenceBuilder;
-
 import org.apache.hyracks.data.std.api.IPointable;
 import org.apache.hyracks.data.std.primitive.BooleanPointable;
 import org.apache.hyracks.data.std.util.ArrayBackedValueStorage;
+import org.apache.vxquery.datamodel.builders.jsonItem.ObjectBuilder;
+import org.apache.vxquery.datamodel.builders.sequence.SequenceBuilder;
 
 public class XDMConstants {
     private static final byte[] BOOLEAN_TRUE_CONSTANT;
```
prestoncarman, line 29: We should probably add JSON_NULL.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/AbstractPointableTest.java
```diff
@@ -0,0 +1,69 @@
...
+        } else if (value instanceof java.lang.Double) {
+            writeDouble((Double) value, abvsInput.getDataOutput());
+        } else if (value instanceof java.lang.String) {
+            writeString((String) value, abvsInput.getDataOutput());
+        }
+        result.set(abvsInput.getByteArray(), start, abvsInput.getLength() - start);
+    }
+
+    private void writeInteger(Integer value, DataOutput dOut) throws IOException {
+        dOut.write(ValueTag.XS_LONG_TAG);
```
prestoncarman, line 50: What is the tag for an Integer?

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+        tvp.set(abvsResult);
+
+        // Check results.
+        if (tvp.getTag() != ValueTag.OBJECT_TAG) {
+            Assert.fail("Type tag is incorrect. Expected: " + ValueTag.OBJECT_TAG + " Got: " + tvp.getTag());
+        }
+        tvp.getValue(op);
+        if (op.getEntryCount() != 0) {
+            Assert.fail("Sequence size is incorrect. Expected: 0 Got: " + op.getEntryCount());
```
prestoncarman, line 77: Update comments to reflect object, not sequence. (throughout file)

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,195 @@
...
+        tvp.set(abvsResult);
+
+        // Check results.
+        if (tvp.getTag() != ValueTag.OBJECT_TAG) {
+            Assert.fail("Type tag is incorrect. Expected: " + ValueTag.OBJECT_TAG + " Got: " + tvp.getTag());
+        }
+
+        tvp.getValue(op);
+        try {
+            op.getKeys(tvp);
```
prestoncarman, line 103: Why get the keys? Is this test for keys or a single field look up?

If you want to test both keys and fields. Start first by checking the object size. Then continue with the rest of the checks.

prestoncarman, line 103: I think this should be a check to see how many fields are in the object.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,195 @@
...
+        } catch (SystemException e) {
+            Assert.fail("Test failed to write the object pointable.");
+        }
+
+        if (tvp.getTag() != ValueTag.XS_STRING_TAG) {
```
prestoncarman, line 108: Do these tests pass? Is this working code?

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+        sp.getEntry(2, tvp);
+        if (!comparePointable(tvp, tvp5)) {
+            Assert.fail(
+                    "Sequence item three is incorrect. Expected: " + ValueTag.XS_STRING_TAG + " Got: " + tvp.getTag());
+        }
+
+        //Test values
+        op.getValue(tvp1, tvp);
+        if (!comparePointable(tvp, tvp2)) {
```
prestoncarman, line 182: Add the key to the failure message.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,195 @@
...
+        if (tvp.getTag() != ValueTag.SEQUENCE_TAG) {
+            Assert.fail("Sequence tag is incorrect. Expected: " + ValueTag.SEQUENCE_TAG + " Got: " + tvp.getTag());
+        }
+        tvp.getValue(sp);
+        if (sp.getEntryCount() != 3) {
+            Assert.fail("Sequence size is incorrect. Expected: 3 Got: " + sp.getEntryCount());
+        }
+        sp.getEntry(0, tvp);
+        if (!comparePointable(tvp, tvp1)) {
+            Assert.fail("Sequence item one is incorrect. Expected: " + ValueTag.XS_LONG_TAG + " Got: " + tvp.getTag());
```
prestoncarman, line 167: Seems this should be checking keys, not values.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.vxquery.datamodel.builders.jsonItem.ObjectBuilder;
+import org.apache.vxquery.datamodel.values.ValueTag;
+import org.apache.vxquery.datamodel.values.XDMConstants;
+import org.apache.vxquery.exceptions.SystemException;
+import org.junit.Test;
+
+import junit.framework.Assert;
+
+public class ObjectByteTest extends AbstractPointableTest {
+    ArrayBackedValueStorage abvsResult = new ArrayBackedValueStorage();
```
prestoncarman, line 32: Please add a test for when items do not exists in the object.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/AbstractPointableTest.java
riyafa, line 50: When you call the getTaggedValuePointable() method with and integer it takes it as instanceof java.lang.Integer and if you call it with a long (a long has a 'l' at the end like '100l') it takes it as instanceof java.lang.Long. When you used integer with this method in the SequenceByteTest class it had only 0s and not the integer because it is not recognized as a Long--I debugged the test to find that

westmann, line 50: Maybe there's some confusion between the XS_INT_TAG and the XS_INTEGER_TAG? 
The XS_INTEGER_TAG should theoretically represent an arbitrary precision integer (but it is represented as a long in VXQuery). 
However, for java.lang.Integer in this test it seems that we should use an XS_INT_TAG and use DataOutput.writeInt to write the number.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
riyafa, line 108: Yes the tests pass

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
riyafa, line 167: It checks both

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
riyafa, line 35: I think we have to have that because otherwise we have to explicitly insert a ValueTag into the byte array when returning keys. If it is already there in the array, we have to only point to it.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/XDMConstants.java
riyafa, line 29: What should be the valueTag for NULL? I could return that if a value is not found for a given key.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
riyafa, line 32: Once the null has been defined I could return null if item do not exist and test that.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/ValueTag.java
riyafa, line 85: I thought it wouldn't be necessary because it is similar to ELEMENT_NODE_TAG, ATTRIBUTE_NODE_TAG etc.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/XDMConstants.java
riyafa, line 29: it should be JS_NULL_TAG right?

prestoncarman, line 29: Yes, it should be the next atomic tag number (I think its 52).

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/ValueTag.java
westmann, line 85: Agreed

PR #39, opened 2016-05-17
Array Builder and Array Pointable

pavlopoul:


prestoncarman:
Your branch has changes from my test branch that were not merged with master. Please remove the extra commits. Only include your change in the pull request.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/builders/jsonItem/ArrayBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.DataOutput;
+import java.io.IOException;
+
+import org.apache.hyracks.data.std.api.IMutableValueStorage;
+import org.apache.hyracks.data.std.api.IValueReference;
+import org.apache.hyracks.data.std.util.ArrayBackedValueStorage;
+import org.apache.vxquery.datamodel.values.ValueTag;
+import org.apache.vxquery.util.GrowableIntArray;
+
+public class ArrayBuilder {
```
prestoncarman, line 12: It seems like ArrayBuilder, ObjectBuilder and SequenceBuilder could all share an abstract class. Do you think this would be useful? I guess it would only hold reset. You might be able to create a version of finish that could be shared, between array and object builder.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/XDMConstants.java
```diff
@@ -52,6 +55,14 @@
...
         }
         EMPTY_SEQUENCE = Arrays.copyOf(abvs.getByteArray(), abvs.getLength());
+        ArrayBuilder ab = new ArrayBuilder();
+        ab.reset(abvs);
+        try {
+            ab.finish();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        EMPTY_ARRAY=Arrays.copyOf(abvs.getByteArray(), abvs.getLength());
```
prestoncarman, line 65: Format code. (Make sure all file are using our code style guidelines.)

prestoncarman, line 65: Remove constant. Its not needed at this point.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ArrayPointable.java
```diff
@@ -0,0 +1,59 @@
+package org.apache.vxquery.datamodel.accessors.jsonItem;
+
+import org.apache.hyracks.api.dataflow.value.ITypeTraits;
+import org.apache.hyracks.data.std.api.IPointable;
+import org.apache.hyracks.data.std.api.IPointableFactory;
+import org.apache.hyracks.data.std.primitive.IntegerPointable;
+import org.apache.hyracks.data.std.primitive.VoidPointable;
+import org.apache.vxquery.datamodel.accessors.SequencePointable;
+
+public class ArrayPointable extends SequencePointable {
```
prestoncarman, line 10: It looks like you extend the SequencePointable, but rewrite all the functions? What is not being overridden? Maybe it should just be separate.

PR #40, opened 2016-05-17
The object datamodel implementation

riyafa:
Previous pull request has conflicts with the master so I have made another pull request


vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+            sb.finish();
+            result.set(abvs);
+        } catch (IOException e) {
+            throw new SystemException(ErrorCode.SYSE0001);
+        }
+    }
+
+    public void getValue(TaggedValuePointable key, TaggedValuePointable result) {
+        int dataAreaOffset = getDataAreaOffset(bytes, start);
```
prestoncarman, line 97: Return a boolean value for if the value is found.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/builders/jsonItem/ObjectBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.IOException;
+
+import org.apache.hyracks.data.std.api.IMutableValueStorage;
+import org.apache.hyracks.data.std.api.IValueReference;
+import org.apache.hyracks.data.std.util.ArrayBackedValueStorage;
+import org.apache.vxquery.datamodel.values.ValueTag;
+import org.apache.vxquery.util.GrowableIntArray;
+
+public class ObjectBuilder {
+    private final GrowableIntArray slots = new GrowableIntArray();
```
prestoncarman, line 32: Extend AbstractJsonBuilder.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,184 @@
...
+        try {
+            op.getKeys(tvp);
+        } catch (SystemException e) {
+            Assert.fail("Test failed to write the object pointable.");
+        }
+
+        if (tvp.getTag() != ValueTag.XS_STRING_TAG) {
+            Assert.fail("Type tag is incorrect. Expected: " + ValueTag.XS_STRING_TAG + " Got: " + tvp.getTag());
+        }
+        if (!FunctionHelper.arraysEqual(tvp, tvp1)) {
```
prestoncarman: Give tvp1 names that describe the data it represents: key or value (tvpKey1, etc.).

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,155 @@
...
+            s = dataAreaOffset + getRelativeEntryStartOffset(i);
+            l = getKeyLength(bytes, s);
+            if (FunctionHelper.arraysEqual(bytes, s, l, key.getByteArray(), key.getStartOffset() + 1,
+                    key.getLength() - 1)) {
+                result.set(bytes, s + l, getEntryLength(i) - l);
+                return;
+            }
+        }
+        if (entryCount == 0 || i == entryCount) {
+            XDMConstants.setFalse(result);
```
prestoncarman: Change the method signature to return boolean. True for found, False for not found. Do not set the result pointable when the item is not found.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,155 @@
...
+    public void getKeys(IPointable result) throws SystemException {
+        try {
+            abvs.reset();
+            slots.clear();
+            dataArea.reset();
+            int dataAreaOffset = getDataAreaOffset(bytes, start);
+            int entryCount = getEntryCount();
+            int s;
+            for (int i = 0; i < entryCount; i++) {
+                s = dataAreaOffset + getRelativeEntryStartOffset(i);
```
prestoncarman: As I thought about this more, I think adding a new method to the Sequence builder would be a simpler solution and reduce code duplication. The new method would be "public void addItem(int tagValue, IValueReference p) throws IOException". The method can then create a tagged value sequence of items for you. This will reduce the variables and code in this class. Also the method could be used else where.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,155 @@
...
+            return VoidPointable.TYPE_TRAITS;
+        }
+
+        @Override
+        public IPointable createPointable() {
+            return new ObjectPointable();
+        }
+    };
+    private static final int ENTRY_COUNT_SIZE = IntegerPointable.TYPE_TRAITS.getFixedLength();
+    private static final int SLOT_SIZE = 4;
```
prestoncarman: Same constant for SLOT_SIZE (as ENTRY_COUNT_SIZE).

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,155 @@
...
+                }
+                out.write(dataArea.getByteArray(), dataArea.getStartOffset(), dataArea.getLength());
+            }
+        } else {
+            out.write(dataArea.getByteArray(), dataArea.getStartOffset(), dataArea.getLength());
+        }
+    }
+
+    public void getValue(TaggedValuePointable key, TaggedValuePointable result) throws SystemException {
+        int dataAreaOffset = getDataAreaOffset(bytes, start);
```
prestoncarman, line 97: You can change this function to only accept UTF8StringPointables for the key argument. The caller should handle the tagged value. Also the result is just a IPointable.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/builders/jsonItem/ObjectBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+        slots.clear();
+        dataArea.reset();
+    }
+
+    @Override
+    public int getValueTag() {
+        return ValueTag.OBJECT_TAG;
+    }
+
+    public void addItem(IValueReference key, IValueReference value) throws IOException {
```
prestoncarman: I think the key here should be a UTF8StringPointable. The object only supports keys of type string.

vxquery-core/src/main/java/org/apache/vxquery/types/BuiltinTypeConstants.java
```diff
@@ -72,5 +72,7 @@
     public static final int XS_NMTOKENS_TYPE_ID = 50;
     public static final int XS_ENTITIES_TYPE_ID = 51;
 
-    public static final int BUILTIN_TYPE_COUNT = 52;
+    public static final int JS_NULL_ID = 52;
```
prestoncarman: JS_NULL_TYPE_ID

vxquery-core/src/test/java/org/apache/vxquery/datamodel/AbstractPointableTest.java
```diff
@@ -28,7 +28,9 @@
 
     protected void getTaggedValuePointable(Object value, IPointable result) throws IOException {
```
prestoncarman, line 30: It may be helpful to add an other argument for including the type tag. This way you can create strings without the type tag. "protected void getTaggedValuePointable(Object value, boolean includeTag, IPointable result) throws IOException "

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private static int getSlotArrayOffset(int start) {
+        return start + ENTRY_COUNT_SIZE;
+    }
+
+    private static int getDataAreaOffset(byte[] bytes, int start) {
+        return getSlotArrayOffset(start) + getEntryCount(bytes, start) * SLOT_SIZE;
+    }
+
+    public void getKeys(IPointable result) throws SystemException {
```
prestoncarman: I think this should throw an IOException instead of changing it to SystemException.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
```diff
@@ -0,0 +1,155 @@
...
+                    out.writeInt(slotArray[i]);
+                }
+                out.write(dataArea.getByteArray(), dataArea.getStartOffset(), dataArea.getLength());
+            }
+        } else {
+            out.write(dataArea.getByteArray(), dataArea.getStartOffset(), dataArea.getLength());
+        }
+    }
+
+    public void getValue(TaggedValuePointable key, TaggedValuePointable result) throws SystemException {
```
prestoncarman: What is causing the SystemException?

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,253 @@
...
+            ob.finish();
+        } catch (IOException e) {
+            Assert.fail("Test failed to write the object pointable.");
+        }
+
+        tvp.set(abvsResult);
+        tvp.getValue(op);
+        try {
+            getTaggedValuePointable("key", tvpKey3);
+            op.getValue(tvpKey3, tvpValue3);
```
prestoncarman: // Something like this...
if (op.getValue(tvpKey3, tvpValue3)) {
   // found key
}

prestoncarman: Or ! op.get...

vxquery-core/src/main/java/org/apache/vxquery/datamodel/values/ValueTag.java
```diff
@@ -74,6 +74,8 @@
     public static final int XS_NMTOKENS_TAG = BuiltinTypeConstants.XS_NMTOKENS_TYPE_ID;
     public static final int XS_ENTITIES_TAG = BuiltinTypeConstants.XS_ENTITIES_TYPE_ID;
 
+    public static final int JS_NULL_TAG = BuiltinTypeConstants.JS_NULL_ID;
```
prestoncarman: Can you also add js:null constant to XDMConstants.

```
    JS_NULL_CONSTANT = new byte[1];
    JS_NULL_CONSTANT[0] = ValueTag.JS_NULL_TAG;
```

Along with the setJsNull(IPointable p) method

prestoncarman:
The changes are coming together.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/AbstractPointableTest.java
riyafa, line 30: The objectbuilder removes the type tag, but now requires a UTF8StringPointable to be passed in for key, so it wouldn't be necessary to have a pointable without the tag when testing

prestoncarman, line 30: See my other comments. Object build should not remove the tag, but require a UTF8StringPointable as the argument. Let the calling class ensure a UTF8StringPointable as the key argument.

prestoncarman, line 30: Consider the the Comment Node example. The constructor ensures the argument type is correct before calling the builder. The same is true for objects. The constructor will ensure the key is of type string and only pass in the pointable for the string (without the tag). 

Caller - CommentNodeConstructorScalarEvaluator
Builder - CommentNodeBuilder

These comment are trying to get all notions of tags out of the pointable and builder. (No more -1 +1 backed in)

riyafa, line 30: Thank you. I got it now

riyafa, line 30: I guess same applied for getValue method of ObjectPointable: it takes a pointable for the string without the tag

riyafa, line 30: And the getKeys method of ObjectPointable: should it return a sequence pointable of keys which are string pointables without valuetag?

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonItem/ObjectPointable.java
riyafa: If object pointable ignores the notion of tags, I think this implementation is not necessary

vxquery-core/src/test/java/org/apache/vxquery/datamodel/AbstractPointableTest.java
prestoncarman, line 30: Include the type tag for the result of the getKeys().

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -0,0 +1,261 @@
...
+            if (!op.getValue(tvpKey1, tvp)) {
+                Assert.fail("key in object. Expected: true Got: false");
+            }
+        } catch (IOException e) {
+            Assert.fail("Test failed to write the object pointable.");
+        }
+    }
+
+    public boolean compareKeys(IPointable tvp1, IPointable tvp2) {
+        return FunctionHelper.arraysEqual(tvp1.getByteArray(), tvp1.getStartOffset() + 1, tvp1.getLength() - 1,
```
prestoncarman: Use the getValue method from TaggedValuePointable to get the string pointable.

riyafa: I have removed the method

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -96,7 +95,7 @@ public void testSingleItemObject() {
         if (tvp.getTag() != ValueTag.XS_STRING_TAG) {
             Assert.fail("Type tag is incorrect. Expected: " + ValueTag.XS_STRING_TAG + " Got: " + tvp.getTag());
         }
-        if (!compareKeys(tvp, tvpKey1)) {
+        if (tvpKey1.compareTo(tvp) < 0) {
```
riyafa: There is a mistake. I will fix that

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/SequencePointable.java
```diff
@@ -39,42 +37,44 @@ public IPointable createPointable() {
             return new SequencePointable();
         }
     };
+    private static final int ENTRY_COUNT_SIZE = IntegerPointable.TYPE_TRAITS.getFixedLength();
```
prestoncarman: Why reorder the methods in this class? Try to minimize the changes.

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
riyafa: I am using this method because the pointable returned from getKeys has tag and the key I have created don't have tag. This is to compare both. Why use getValue method of TaggedValuePointable in this case? Sorry about the confusion

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
```diff
@@ -252,4 +253,10 @@ public void testItemNotExistObject() {
             Assert.fail("Test failed to write the object pointable.");
         }
     }
+
+    private boolean compareKeys(IPointable tvp1, IPointable tvp2) {
+
+        return FunctionHelper.arraysEqual(tvp1.getByteArray(), tvp1.getStartOffset() + 1, tvp1.getLength() - 1,
+                tvp2.getByteArray(), tvp2.getStartOffset(), tvp2.getLength());
+    }
```
riyafa, line 263: I am using this method because the pointable returned from getKeys has tag and the key I have created don't have tag. This is to compare both. Why use getValue method of TaggedValuePointable in this case? Sorry about the confusion. It is ok to have +1 and -1 in a test method. Isn't it?

vxquery-core/src/test/java/org/apache/vxquery/datamodel/ObjectByteTest.java
prestoncarman: Something like this...

private IPointable vp = VoidPointable.FACTORY.createPointable();
public boolean compareKeys(TaggedValuePointable tvp1, IPointable tvp2) {
    tvp1.getValue(vp)
    return FunctionHelper.arraysEqual(vp.getByteArray(),
vp.getStartOffset(), vp.getLength(),
        tvp2.getByteArray(), tvp2.getStartOffset(), tvp2.getLength());
}

prestoncarman:
I merged all but the last commit. The last commit was unrelated to the object builder and pointable. Make sure to keep each branch clean with only the changes for that topic/issue.
Please close.

PR #48, opened 2016-05-29
changes to array grammar

pavlopoul:


vxquery-core/src/main/javacc/xquery-grammar.jj
```diff
@@ -1796,6 +1796,7 @@ ASTNode PrimaryExpr():
         | LOOKAHEAD(2) expr = OrderedExpr()
         | LOOKAHEAD(2) expr = UnorderedExpr()
         | expr = Constructor()
+        | expr = JsonConstructor()
```
westmann, line 1799: Could we move the JsonConstructor inside of the Constructor production?

vxquery-core/src/main/javacc/xquery-grammar.jj
```diff
@@ -1937,6 +1938,47 @@ ASTNode FunctionCall()  :
...
+{
+    result = ArrayConstructor()
+    {
+		return result;
+	}
+}
+
+ASTNode ArrayConstructor()  :
+{
+	Token start, end;
```
westmann, line 1954: These Tokens are not really needed, as we don't use them again.

vxquery-core/src/main/javacc/xquery-grammar.jj
```diff
@@ -1937,6 +1938,47 @@ ASTNode FunctionCall()  :
...
+}
+{
+	start="[" items=ArrayItemList() end="]" {
+		ArrayConstructorNode array = new ArrayConstructorNode(createSourceLocation(start));
+		array.setItems(items);
+		return array;
+	}
+}
+
+List<ASTNode> ArrayItemList()  :
```
westmann, line 1965: If we just reuse the expression's ability to capture a sequence, we could make the production of the ArrayConstructor `"[" expr = Expr() "]"` and avoid this ArrayItemList.

PR #53, opened 2016-06-09
Corrections in identation and single item return

pavlopoul:


vxquery-core/src/main/java/org/apache/vxquery/serializer/XMLSerializer.java
```diff
@@ -441,6 +446,34 @@ private void printSequence(PrintStream ps, SequencePointable seqp, String betwee
         }
     }
 
+    private void printArrayNode(PrintStream ps, TaggedValuePointable tvp) {
+        ArrayPointable ap = pp.takeOne(ArrayPointable.class);
+
+        try {
+            tvp.getValue(ap);
+            if (tvp.getTag() == ValueTag.ARRAY_TAG) {
```
prestoncarman, line 454: The TVP is an array. No need to check it in this function. Take a look at he other print functions.

vxquery-core/src/main/java/org/apache/vxquery/serializer/XMLSerializer.java
```diff
@@ -441,6 +446,34 @@ private void printSequence(PrintStream ps, SequencePointable seqp, String betwee
         }
     }
 
+    private void printArrayNode(PrintStream ps, TaggedValuePointable tvp) {
```
prestoncarman, line 449: printArray -- JSON array is not a node.

vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-operators.xml
```diff
@@ -16,1008 +16,1139 @@
   limitations under the License.
 -->
 <operators>
-    <!-- op:add-dayTimeDuration-to-date( $arg1 as xs:date, $arg2 as xs:dayTimeDuration)  as xs:date -->
+    <!-- op:add-dayTimeDuration-to-date( $arg1 as xs:date, $arg2 as xs:dayTimeDuration)
```
prestoncarman, line 19: Please do not change the file format.

PR #56, opened 2016-06-16
Implement JSONiq functions

riyafa:
…rray and null


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/misc/FnDataScalarEvaluatorFactory.java
```diff
@@ -71,10 +71,12 @@ protected void evaluate(TaggedValuePointable[] args, IPointable result) throws S
                     int seqLen = seq.getEntryCount();
                     for (int j = 0; j < seqLen; ++j) {
                         seq.getEntry(j, p);
+                        check(p);
                         ah.atomize(p, ppool, tempTVP);
```
prestoncarman, line 74: Move the check to atomize helper class. This way any user of the atomize helper class will get the new error message.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/cast/AbstractCastToOperation.java
```diff
@@ -226,4 +226,8 @@ public void convertNormalizedString(UTF8StringPointable stringp, DataOutput dOut
     public void convertToken(UTF8StringPointable stringp, DataOutput dOut) throws SystemException, IOException {
         throw new SystemException(ErrorCode.XPTY0004);
     }
+
+    public void convertNull(DataOutput dOut) throws SystemException, IOException {
```
prestoncarman, line 230: Also update CastableAsStringOperation for the null case.

prestoncarman, line 230: Also add a test case for this option.

riyafa, line 230: I think I have already added a testcase for null with fn:string in file fn_string1.xq. What do you mean by "this options"? Does anything change if I add null to CastableAsStringOperation that needs to be tested?

prestoncarman, line 230: Here is an example query: 

xs:string("test") castable as js:null

Does this help to explain? Add a test query with the above syntax and update CastableAsStringOperation similar to CastAsStringOperation.

prestoncarman, line 230: I see you update the code, now we just need a query to test it.

prestoncarman, line 230: Also, add a direct test for casting:
xs:string("test") cast as js:null

Change the query as necessary for your test.

prestoncarman:
One last comment: Why comment out one of the index tests?
Ready to squash commits. If you need to start a new branch and do a squash merge to get a single commit. Remember to make a nice commit message like an e-mail: first line is a title (<50 chars) and line break then the body.

riyafa:
When I comment that test all the tests pass. I will uncomment it and create a pull request with only one commit.

prestoncarman:
Go ahead and leave it commented out if the tests pass.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/cast/CastScalarEvaluatorFactory.java
```diff
@@ -56,9 +56,9 @@ protected void evaluate(TaggedValuePointable tvp, IPointable result) throws Syst
                 int tid = tvp.getTag();
                 try {
                     switch (tid) {
-                    /**
-                     * Primitive Datatypes (Alphabetical)
-                     */
+                        /**
+                         * Primitive Datatypes (Alphabetical)
+                         */
```
westmann, line 61: Please remove the tabs.

westmann:
Looks good to me as well.

prestoncarman:
Merged the squashed version of this change.

PR #60, opened 2016-06-21
VXQUERY-213 Comment out indexing tests

pavlopoul:


prestoncarman:
We can abandon this change.

PR #62, opened 2016-06-22
Completed Update index query implementation

menakaj:
**Update Index Query**
The update-index query takes two arguments, collection directory and index directory. 

Updated 24/06/2016 : update index query only takes index directory as an argument

It shares some of the functionalities from build-index-on-collection query so, some changes were done to the following classes in order to use them in updating index process and to maximize code reuse
- IndexConstructorUtil.java : Created a new function to get an instance of IndexDocumentBuilder which can be used in IndexUpdater.java class.
- IndexDocumentBuilder.java : Added a new string filed containing the corresponding file path which is needed to retrieve a document related to an XML file.

**Metadata handling**
Here a POJO is created to properly manage the metadata for a file. (XmlMetadata.java)
Currently it contains following fields. 
- File path
- File Name (Not used)
- Checksum String

When storing metadata, a HashMap is created with file path as the key and XmlMetadata object. This map is then serialized and written to a file named  _metadata.file_ and stored in the same directory where the index is stored.

**Update Index process**
- If a file is detected as modified, the current index document related to that file is deleted and newly created index document is added. 
- If a new file is detected, a new index document will be created and added to the existing index.
- If the file is deleted, delete the index document related to that file.
- After every task, update the metadata object and after all processes completed, write the new metadata map to the file. 

Please review the pull request and merge.

Thank you 
Menaka


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/index/IndexConstructorUtil.java
```diff
@@ -111,25 +127,41 @@ public static void evaluate(TaggedValuePointable[] args, IPointable result, UTF8
...
-                XMLParser parser = new XMLParser(false, nodeIdProvider, nodeId);
-                FunctionHelper.readInDocFromString(file.getPath(), bbis, di, abvsFileNode, parser);
-
-                nodep.set(abvsFileNode.getByteArray(), abvsFileNode.getStartOffset(), abvsFileNode.getLength());
-
-                //Add the document to the index
-                //Creates one lucene doc per file
-                IndexDocumentBuilder ibuilder = new IndexDocumentBuilder(nodep, writer);
+//                XMLParser parser = new XMLParser(false, nodeIdProvider, nodeId);
+//                FunctionHelper.readInDocFromString(file.getPath(), bbis, di, abvsFileNode, parser);
```
prestoncarman: Remove commented out code.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/index/IndexConstructorUtil.java
```diff
@@ -111,25 +127,41 @@ public static void evaluate(TaggedValuePointable[] args, IPointable result, UTF8
...
                 ibuilder.printStart();
 
+                if (!isMetaFilePresent) {
+                    XmlMetadata xmlMetadata = new XmlMetadata();
+                    xmlMetadata.setPath(file.getCanonicalPath());
+                    xmlMetadata.setFileName(file.getName());
+                    try {
+                        xmlMetadata.setMd5(metaFileUtil.generateMD5(file));
+                    } catch (NoSuchAlgorithmException e) {
+                        e.printStackTrace();
```
prestoncarman: Please don't swallow exceptions. One option would be to wrap the exception into a SystemException with error code SYSE0001.

prestoncarman:
Please add tests for your new functionality.

PR #64, opened 2016-06-22
Updated the rat configuration along with a fix for unit testing on jenkins.

prestoncarman:
The rat report is now only generated for the top level project. 
HDFS now closes it files when creating the HDFS data.

### DIFF
--- a/vxquery-core/pom.xml
+++ b/vxquery-core/pom.xml
@@ -121,12 +121,12 @@
                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
                 </configuration>
             </plugin>
-            <!-- <plugin> <groupId>org.apache.maven.plugins</groupId> <artifactId>maven-site-plugin</artifactId> 
+            <!-- <plugin> <groupId>org.apache.maven.plugins</groupId> <artifactId>maven-site-plugin</artifactId>
                 </plugin> -->
         </plugins>
         <pluginManagement>
             <plugins>
-                <!--This plugin's configuration is used to store Eclipse m2e settings 
+                <!--This plugin's configuration is used to store Eclipse m2e settings
                     only. It has no influence on the Maven build itself. -->
                 <plugin>
                     <groupId>org.eclipse.m2e</groupId>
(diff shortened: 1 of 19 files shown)


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/util/FunctionHelper.java
```diff
@@ -1234,18 +1234,17 @@ public static void readInDocFromString(String fName, ByteBufferInputStream bbis,
...
-                } catch (FileNotFoundException e) {
-                    // TODO Auto-generated catch block
-                    System.err.println(e);
+                    fs.close();
                 } catch (IOException e) {
-                    // TODO Auto-generated catch block
-                    System.err.println(e);
+                    throw new HyracksDataException(e);
+                } finally {
                 }
```
westmann, line 1254: Why do we have an empty finally block?

vxquery-xtest/pom.xml
```diff
@@ -199,20 +199,6 @@
...
-                <configuration>
-                    <excludes>
-                        <exclude>**/ExpectedTestResults/**/*</exclude>
-                        <exclude>**/xqts.txt</exclude>
-                        <exclude>test-suite*/**/*</exclude>
-                    </excludes>
-                    <numUnapprovedLicenses>0</numUnapprovedLicenses>
-                    <maxmemory>2g</maxmemory>
-                </configuration>
-            </plugin>
```
westmann, line 214: Why do we remove rat here?

vxquery-core/pom.xml
```diff
@@ -341,24 +341,11 @@
...
-                <version>${apache-rat-plugin.version}</version>
-                <configuration>
-                    <excludes>
-                        <exclude>testsuites/**/*</exclude>
-                        <exclude>reports/**/*</exclude>
-                    </excludes>
-                    <numUnapprovedLicenses>0</numUnapprovedLicenses>
-                    <maxmemory>2g</maxmemory>
-                </configuration>
-            </plugin>
```
westmann, line 360: Why do we remove rat here?

vxquery-xtest/pom.xml
prestoncarman, line 214: The exclude statements were getting out of sync with the top level project. Do you know a way to keep them in-sync. Otherwise we can just remove the rat report check from lower level projects and keep it at the top level.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/util/FunctionHelper.java
prestoncarman, line 1254: Ops, I will fix that.

PR #66, opened 2016-06-23
JSONiq array navigation functions and testing

pavlopoul:


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/misc/FnSizeScalarEvaluatorFactory.java
```diff
@@ -0,0 +1,77 @@
...
+        final ArrayPointable ap = (ArrayPointable) ArrayPointable.FACTORY.createPointable();
+
+        public FnSizeScalarEvaluator(IScalarEvaluator[] args) {
+            super(args);
+        }
+
+        @Override
+        protected void evaluate(TaggedValuePointable[] args, IPointable result) throws SystemException {
+            TaggedValuePointable tvp = args[0];
+            if (!(tvp.getTag() != ValueTag.ARRAY_TAG || tvp.getTag() != ValueTag.OBJECT_TAG)) {
```
prestoncarman, line 61: Is this for objects also?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/ValueScalarEvaluator.java
```diff
@@ -0,0 +1,86 @@
...
+            if (tvp2.getTag() != ValueTag.XS_INTEGER_TAG) {
+                throw new SystemException(ErrorCode.FORG0006);
+            }
+            TaggedValuePointable tempTvp = ppool.takeOne(TaggedValuePointable.class);
+            mvs.reset();
+            try {
+                ab.reset(mvs);
+                tvp1.getValue(ap);
+                tvp2.getValue(lp);
+                if ((int) lp.getLong() > ap.getEntryCount()) {
```
prestoncarman, line 69: Return an empty sequence when the index is not found. From the spec:
If $i is comprised between 1 and jdm:size($a), then $a($i) returns the value of the pair at position $i, i.e., jdm:value($a, $i).
Otherwise (i.e., if $i is an invalid position), an empty sequence is returned.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/misc/FnSizeScalarEvaluatorFactory.java
pavlopoul, line 61: yes....I don't think that objects need something different from that.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/ValueScalarEvaluator.java
pavlopoul, line 69: I thought that since I don't add an item in the sequence builder, I return an empty sequence. Should I do something else?

PR #70, opened 2016-06-23
JSONiq size function, array navigation and tests

pavlopoul:


prestoncarman:
I updated master with several changes and this change seems to be out of sync now. Please update.

PR #75, opened 2016-06-30
Implement object navigation and the functions jn:keys and jn:null wit…

riyafa:
…h test cases


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnKeysScalarEvaluator.java
```diff
@@ -0,0 +1,119 @@
...
+        sb = new SequenceBuilder();
+        abvs = new ArrayBackedValueStorage();
+    }
+
+    @Override
+    protected void evaluate(TaggedValuePointable[] args, IPointable result) throws SystemException {
+        TaggedValuePointable tvp1 = args[0];
+        ObjectPointable op;
+        if (tvp1.getTag() == ValueTag.SEQUENCE_TAG) {
+            List<TaggedValuePointable> pointables = new ArrayList<>();
```
prestoncarman: Move the creation of the pointables object to the constructor. In the evaluate, just reset or clear the pointables variable.

prestoncarman:
Reformat your commit message to have a title (line 1) less then 50 characters, followed by a line break and complete details of the commit. Similar to an e-mail subject and body.

PR #76, opened 2016-06-30
Implement object navigation

riyafa:
Implement jdm:keys, jn:keys and jn:null and extend jdm:values with test cases


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnKeysScalarEvaluator.java
```diff
@@ -0,0 +1,120 @@
...
+        sb = new SequenceBuilder();
+        abvs = new ArrayBackedValueStorage();
+    }
+
+    @Override
+    protected void evaluate(TaggedValuePointable[] args, IPointable result) throws SystemException {
+        TaggedValuePointable tvp1 = args[0];
+        ObjectPointable op;
+        if (tvp1.getTag() == ValueTag.SEQUENCE_TAG) {
+            pointables = new ArrayList<>();
```
prestoncarman: Partially correct. Also move the instantiation of the array outside the evaluate. You can used pointables.clear() inside evaluate.

PR #77, opened 2016-07-01
VXQUERY-215 Left operand of a pair is ExprSingle

riyafa:


vxquery-xtest/src/test/resources/Queries/XQuery/Json/Object/q11_object.xq
```diff
@@ -0,0 +1,22 @@
...
+   software distributed under the License is distributed on an
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+   specific language governing permissions and limitations
+   under the License. :)
+
+(: Json Object Query :)
+(: Issue VXQUERY-215 :)
+{
+    <Name>Joe</Name> : "Joe"
```
prestoncarman: I think this is a new test ... based on my comment. Not sure. Here is the query I was thinking about: 

```
{
    <Name>Joe</Name> : "Joe",
    "Joe" : "Joe"
}
```

PR #80, opened 2016-07-02
Adjust jn:null, jdm:keys, newlines at end of tests

riyafa:
Based on post commit comments


vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-functions.xml
```diff
@@ -1158,7 +1158,7 @@
 
     <!-- jn:null() as js:null -->
     <function name="jn:null">
-        <return type="xs:string*"/>
+        <return type="js:null*"/>
```
westmann: Should just be "js:null" as it always returns exactly one null value.

westmann:
Looks good to me.

vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-operators.xml
```diff
@@ -655,9 +655,10 @@
         <runtime type="scalar" class="org.apache.vxquery.runtime.functions.json.ValueScalarEvaluatorFactory"/>
     </operator>
 
+    jdm:keys($o as object()) as xs:string*
```
prestoncarman: This should be a xml comment.

prestoncarman:
Merged squashed version of this change.

PR #81, opened 2016-07-02
Implementation of json Parser and test cases

pavlopoul:


vxquery-core/src/main/java/org/apache/vxquery/compiler/rewriter/rules/ReplaceSourceMapInDocExpression.java
```diff
@@ -102,6 +102,14 @@ public boolean rewritePost(Mutable<ILogicalOperator> opRef, IOptimizationContext
                     modified = true;
                 }
             }
+            docExpression = ExpressionToolbox.findFirstFunctionExpression(expression,
+                    BuiltinFunctions.FN_DOC_JSON_1.getFunctionIdentifier());
```
westmann, line 106: Should be called jn:json-doc [1]

[1] http://jsoniq.org/docs/JSONiqExtensionToXQuery/html-single/index.html#idm139680622843744

vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-functions.xml
```diff
@@ -323,6 +323,13 @@
         <return type="document-node()?"/>
         <runtime type="scalar" class="org.apache.vxquery.runtime.functions.node.FnDocScalarEvaluatorFactory"/>
     </function>
+    
+    <!-- jn:doc($uri as  xs:string?) as document-node()?  -->
+    <function name="fn:doc-json">
```
westmann, line 328: Should be called jn:json-doc [1]

[1] http://jsoniq.org/docs/JSONiqExtensionToXQuery/html-single/index.html#idm139680622843744

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+        DataOutput outResult = result.getDataOutput();
+        DataOutput out = atomic.getDataOutput();
+        StringValueBuilder svb = new StringValueBuilder();
+        String lastItem = null;
+        String startItem = null;
+        boolean nested = false;
+        JsonParser parser = factory.createParser(file);
+        while (!parser.isClosed()) {
+            JsonToken token = parser.nextToken();
+            if (token == JsonToken.START_ARRAY) {
```
westmann, line 65: As JsonToken is an enum, this should be a switch statement.

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+
+    final JsonFactory factory;
+    protected final ArrayBackedValueStorage atomic;
+    protected final List<ArrayBuilder> abStack;
+    protected final List<ObjectBuilder> obStack;
+    protected final List<ArrayBackedValueStorage> abvsStack;
+    protected final List<UTF8StringPointable> spStack;
+    protected final List<String> itemStack;
+
+    public JSONParser() throws JsonParseException, IOException {
```
westmann, line 44: No IOException thrown here.

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+        boolean nested = false;
+        JsonParser parser = factory.createParser(file);
+        while (!parser.isClosed()) {
+            JsonToken token = parser.nextToken();
+            if (token == JsonToken.START_ARRAY) {
+                if (!nested) {
+                    abvsStack.add(new ArrayBackedValueStorage());
+                    abStack.add(new ArrayBuilder());
+                }
+                itemStack.add("array");
```
westmann, line 70: These values ("array", "object"), should be enums (or ints).

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+                }
+            }
+            if (token == JsonToken.END_ARRAY) {
+                if (itemStack.size() == 1) {
+                    abStack.get(abStack.size() - 1).finish();
+                    abStack.remove(abStack.size() - 1);
+                    itemStack.remove(itemStack.size() - 1);
+                    nested = true;
+                    lastItem = "array";
+                } else if (itemStack.get(itemStack.size() - 2) == "array") {
```
westmann, line 131: Comparing using == compares for object identity, not for string equality. It works in this case, as the strings are literals in the java source code and thus the same string will be represented by the same object. However, this is risky code. Using an enum instead of a string solves this problem as well.

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+                startItem = "object";
+            }
+        }
+        if (startItem == "array" || startItem == "object") {
+            outResult.write(abvsStack.get(0).getByteArray());
+        } else {
+            outResult.write(atomic.getByteArray());
+        }
+    }
+}
```
westmann, line 188: add a new line at the end of the file

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+        spStack = new ArrayList<UTF8StringPointable>();
+        itemStack = new ArrayList<String>();
+    }
+
+    public void parseDocument(File file, ArrayBackedValueStorage result)
+            throws NumberFormatException, JsonParseException, IOException {
+        DataOutput outResult = result.getDataOutput();
+        DataOutput out = atomic.getDataOutput();
+        StringValueBuilder svb = new StringValueBuilder();
+        String lastItem = null;
```
westmann, line 59: Isn't lastItem always the top of the itemStack?

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+                    obStack.add(new ObjectBuilder());
+                }
+                itemStack.add("object");
+                lastItem = "object";
+                abvsStack.get(abvsStack.size() - 1).reset();
+                obStack.get(obStack.size() - 1).reset(abvsStack.get(abvsStack.size() - 1));
+            }
+            if (token == JsonToken.FIELD_NAME) {
+                UTF8StringPointable sp = new UTF8StringPointable();
+                ArrayBackedValueStorage key = new ArrayBackedValueStorage();
```
westmann, line 87: It seems that we should not need a new UTF8StringPointable and a new ArrayBackedValueStorage for each field name. Instead we should only need one per level of the depth of the JSON object. So that maximum number of UTF8StringPointables and ArrayBackedValueStorages that should be created is the nesting depth of JSON items.

More generally, it seems useful for all objects that we're allocating to maintain a stack of these objects that only grows up to the maximal depth and that gets garbage collected at the end of the complete parsing process. To reuse those objects we just need to a) maintain the current depth that we're parsing at to find the object in the Stack and b) reset() the object before reusing it.

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+                }
+
+                startItem = "array";
+
+            }
+            if (token == JsonToken.END_OBJECT) {
+                if (itemStack.size() == 1) {
+                    obStack.get(obStack.size() - 1).finish();
+                    obStack.remove(obStack.size() - 1);
+                    spStack.remove(spStack.size() - 1);
```
westmann, line 156: There's quite a bit of redundancy in these cases (also for the END_ARRAY case), e.g. the first 3 lines. It would be nice to extract that and have the code only once.

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+                    itemStack.remove(itemStack.size() - 1);
+                    nested = true;
+                    lastItem = "object";
+                } else if (itemStack.get(itemStack.size() - 2) == "object") {
```
westmann, line 160: itemStack.get(itemStack.size() - 2) is accessed twice here. It would be good to create a local variable for it.

vxquery-core/src/main/java/org/apache/vxquery/jsonparser/JSONParser.java
```diff
@@ -0,0 +1,188 @@
...
+                spStack.add(sp);
+                DataOutput outk = key.getDataOutput();
+                svb.write(parser.getText(), outk);
+                spStack.get(spStack.size() - 1).set(key);
+            }
+            if (token == JsonToken.VALUE_NUMBER_INT) {
+                atomic.reset();
+                out.write(ValueTag.XS_INTEGER_TAG);
+                out.writeLong(parser.getLongValue());
+                if (lastItem == "array") {
```
westmann, line 98: We could create a method for this if-then-else that could be called for all atomic values.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnDocScalarEvaluatorFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+            @Override
+            protected void evaluate(TaggedValuePointable[] args, IPointable result) throws SystemException {
+                TaggedValuePointable tvp = args[0];
+                if (tvp.getTag() != ValueTag.XS_STRING_TAG) {
+                    throw new SystemException(ErrorCode.FORG0006);
+                }
+                tvp.getValue(stringp);
+                try {
+                    JSONParser parser = new JSONParser();
```
westmann, line 66: It would be nice, if we could build the parser in a way that in can be reset and re-used instead of creating a new one for each function evaluation.

vxquery-xtest/src/test/resources/TestSources/json/atomic_double.json
```diff
@@ -0,0 +1 @@
+1.2
```
westmann, line 1: Add a new line at the end of the file.

vxquery-xtest/src/test/resources/TestSources/json/atomic_int.json
```diff
@@ -0,0 +1 @@
+1
```
westmann, line 1: Add a new line at the end of the file.

vxquery-xtest/src/test/resources/TestSources/json/atomic_string.json
```diff
@@ -0,0 +1 @@
+"string"
```
westmann, line 1: Add a new line at the end of the file.

vxquery-xtest/src/test/resources/cat/SingleQuery.xml
```diff
@@ -23,4 +23,9 @@
       <query name="add" date="2014-08-18"/>
       <output-file compare="Text">add.txt</output-file>
    </test-case>
+   <test-case name="simple-fn-doc" FilePath="Simple/" Creator="Preston Carman">
+      <description>Adds two numbers.</description>
+      <query name="fn_doc" date="2014-08-18"/>
+      <output-file compare="Text">fn_doc.txt</output-file>
+   </test-case>
 </test-group>
```
westmann, line 31: Add a new line at the end of the file.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnDocScalarEvaluatorFactory.java
```diff
@@ -0,0 +1,77 @@
...
+                    FunctionHelper.readInJnDocFromPointable(stringp, bbis, di, abvs, parser);
+                } catch (IOException e) {
+                    e.printStackTrace();
```
westmann, line 69: Please throw an exception with error code FODC0002 here.

PR #86, opened 2016-07-06
VXQUERY-211 Support "?:" in object constructor

riyafa:


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/jsonitem/ObjectConstructorScalarEvaluator.java
```diff
@@ -64,37 +73,54 @@ private boolean isDuplicate(TaggedValuePointable tempKey) {
...
+                value = args[i + 1];
+                qmc = args[i + 2];
+                if (!isDuplicate(key)) {
+                    pointables[i / 3] = (TaggedValuePointable) TaggedValuePointable.FACTORY.createPointable();
+                    key.getValue(pointables[i / 3]);
+                    sp.set(vp);
+                    if (value.getTag() == ValueTag.SEQUENCE_TAG) {
+                        qmc.getValue(bp);
+                        value.getValue(seqp);
+                        if (bp.getBoolean()) {
```
prestoncarman: What about this case?

{"test" ?: (1,2,3)}

westmann: Should the check on bp be after the check for the empty sequence in line 97?

It would indeed be good to have a test case that tests a non-empty sequence after a "?:".

prestoncarman:
Looks good. Please rebase and squash commit.

PR #87, opened 2016-07-06
VXQUERY-221: Removing some extra exception wrapping.

prestoncarman:
I found these places, based on VXQUERY-221, where we can clean up the SystemExceptions.


westmann:
Looks good to me

PR #94, opened 2016-07-12
Array unboxing and testing

pavlopoul:


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/KeysOrMembersScalarEvaluator.java
```diff
@@ -52,6 +64,27 @@ protected void evaluate(TaggedValuePointable[] args, IPointable result) throws S
                 throw new SystemException(ErrorCode.SYSE0001, e);
 
             }
+        } else if (tvp1.getTag() == ValueTag.ARRAY_TAG) {
+            abvs.reset();
+            sb.reset(abvs);
+            TaggedValuePointable tempTvp = (TaggedValuePointable) TaggedValuePointable.FACTORY.createPointable();
```
prestoncarman: Reuse pointable for each evaluation call.

PR #97, opened 2016-07-14
some array access refactoring

westmann:
- extract common code in SequencePointable and ArrayPointable to
  AbstractSequencePointable
- add code to extract sequences from arrays to ArrayPointable and use it
  in JnMembersScalarEvaluator and KeysOrMembersScalarEvaluator
- reduce complexity of KeysOrMembersScalarEvaluator.evaluate by introducing
  a switch and pulling exception wrapping further out

### DIFF
--- a/vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonitem/ArrayPointable.java
+++ b/vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonitem/ArrayPointable.java
@@ -16,16 +16,17 @@
  */
 package org.apache.vxquery.datamodel.accessors.jsonitem;
 
+import java.io.IOException;
+
 import org.apache.hyracks.api.dataflow.value.ITypeTraits;
-import org.apache.hyracks.data.std.api.AbstractPointable;
 import org.apache.hyracks.data.std.api.IPointable;
 import org.apache.hyracks.data.std.api.IPointableFactory;
-import org.apache.hyracks.data.std.primitive.IntegerPointable;
 import org.apache.hyracks.data.std.primitive.VoidPointable;
+import org.apache.vxquery.datamodel.accessors.AbstractSequencePointable;
+import org.apache.vxquery.datamodel.accessors.TaggedValuePointable;
+import org.apache.vxquery.datamodel.builders.sequence.SequenceBuilder;
 
-public class ArrayPointable extends AbstractPointable {
-    private static final int ENTRY_COUNT_SIZE = IntegerPointable.TYPE_TRAITS.getFixedLength();
-    private static final int SLOT_SIZE = IntegerPointable.TYPE_TRAITS.getFixedLength();
+public class ArrayPointable extends AbstractSequencePointable {
     public static final IPointableFactory FACTORY = new IPointableFactory() {
         private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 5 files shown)


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/KeysOrMembersScalarEvaluator.java
```diff
@@ -48,45 +46,31 @@ public KeysOrMembersScalarEvaluator(IHyracksTaskContext ctx, IScalarEvaluator[]
...
-            } catch (IOException e) {
-                throw new SystemException(ErrorCode.SYSE0001, e);
-            }
-            result.set(abvs);
+        } catch (IOException e) {
+            throw new SystemException(ErrorCode.SYSE0001, e);
         }
     }
-
 }
```
pavlopoul, line 76: This change looks great...the code is simpler now

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnMembersScalarEvaluator.java
```diff
@@ -34,68 +34,50 @@
...
+                int size1 = sp.getEntryCount();
                 for (int i = 0; i < size1; i++) {
-                    sp1.getEntry(i, tempTvp);
+                    sp.getEntry(i, tempTvp);
                     if (tempTvp.getTag() == ValueTag.ARRAY_TAG) {
-                        membersSequence(tempTvp, result, tvp1);
+                        tempTvp.getValue(ap);
+                        ap.appendSequence(sb);
                     } else {
                         XDMConstants.setEmptySequence(result);
```
prestoncarman: I don't think this is correct. Setting the empty sequence here is useless. When the sequence is saved to the result this value is overwritten. I think the else can be removed.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnMembersScalarEvaluator.java
```diff
@@ -34,68 +34,50 @@
...
                     }
                 }
-            } finally {
-                ppool.giveBack(tempTvp);
+            } else if (tvp.getTag() == ValueTag.ARRAY_TAG) {
+                tvp.getValue(ap);
+                ap.appendSequence(sb);
+            } else {
+                XDMConstants.setEmptySequence(result);
```
prestoncarman: Same comment here about this empty sequence. I think the code worked because the sequence builder actually adds the empty sequence.

vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/jsonitem/ArrayPointable.java
```diff
@@ -40,36 +41,13 @@ public IPointable createPointable() {
...
-        return getEntryOffsetValue(bytes, start, idx) - getRelativeEntryStartOffset(idx);
-    }
-
-    private static int getOffsetsStart(int start) {
-        return start + ENTRY_COUNT_SIZE;
-    }
-
-    private static int getDataStart(byte[] bytes, int start) {
-        return getOffsetsStart(start) + getEntryCount(bytes, start) * SLOT_SIZE;
+    public void appendSequence(SequenceBuilder sb) throws IOException {
```
prestoncarman: Should the title be different? You are not appending a sequence, but all the individual items in the sequence.

westmann:
@ecarm002: addressed your comments

prestoncarman:
Thanks, looks good.

PR #100, opened 2016-07-15
Implement SimpleObjectUnionConstructor

riyafa:


vxquery-core/src/main/javacc/xquery-grammar.jj
```diff
@@ -1999,8 +1999,9 @@ ASTNode JsonConstructor() :
 }
 {
     (
-        result = ObjectConstructor()
-        | result = ArrayConstructor()
+        result = ArrayConstructor()
+        | LOOKAHEAD(2) result = ObjectConstructor()
```
westmann: Do we need the `LOOKAHEAD(2)` here? Shouldn't `{` be a different token than `{|`?

prestoncarman:
Looks good to me. Go ahead and prepare for merge.

PR #105, opened 2016-07-19
Working version of libjn:descendant-arrays

pavlopoul:


vxquery-xtest/src/test/resources/Queries/XQuery/Json/Libraries/descendant_arrays.xq
```diff
@@ -0,0 +1,24 @@
...
+   Unless required by applicable law or agreed to in writing,
+   software distributed under the License is distributed on an
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+   specific language governing permissions and limitations
+   under the License. :)
+
+(: JSONiq libjn:descendant_arrays :)
+(
+    libjn:descendant-arrays(({ "foo" : { "bar" : [ 1, 2 ] } },
```
westmann, line 20: Could you also add a test with more than one array at the same level of nesting?

westmann:
Something is confusing here. It seems that this change includes the refactoring that I did a few days ago. Also, there seem to be conflicts with master. Could you resolve the conflicts and make sure, that only your new code is in the change?

PR #112, opened 2016-07-20
Update testing for XQTS test special case.

prestoncarman:
Altered the test result checking process to allow for XQTS tests which have either an error or a result.


westmann:
Looks good to me - haven't run tests ...

PR #115, opened 2016-07-23
VXQUERY-64 fn:trace is now supported

pavlopoul:


prestoncarman:
Looks good to me. Please prepare for the merge (single commit with a nice message).

PR #117, opened 2016-07-26
Changes to arithmetic operation semantics

riyafa:


westmann:
Looks good to me.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/util/ArithmeticHelper.java
```diff
@@ -98,6 +99,8 @@ public void compute(TaggedValuePointable tvp1, TaggedValuePointable tvp2, IPoint
                     doublep1.set(abvsArgument1.getByteArray(), abvsArgument1.getStartOffset() + 1,
                             DoublePointable.TYPE_TRAITS.getFixedLength());
                     break;
+                case ValueTag.JS_NULL_TAG:
```
prestoncarman: It seems a little out of place to have the Null type in a switch statement that is used to convert types.

prestoncarman:
Looks good. Prepare for merge.

PR #121, opened 2016-07-26
Issue VXQUERY-225-fn:serialize

pavlopoul:


westmann:
Just a few quick thoughts/questions:
1. The escaping of the XML entities in the tests looks ugly, but it might be the right thing to do - haven't thought carefully about this. 
2. Where do those entities get introduced? During the evaluation of fn:serialize or in the serialization of the final output?
3. If it's the serialization of the final output, do we have a way to avoid that and get a "readable" output for some queries (e.g. by wrapping the strings in some container)?
4. Do the XQTS tests for fn:serialize pass?

prestoncarman:
1. It seems like escaping XML entities is nice for some scenarios, but I am not sure if its right for all.
2. It seems the XML entities are added when serialization of the final output. (At least that is my guess from looking at code. Christina can you confirm with your new trace function?)
3. Nothing currently exists.
4. The fn:serialize function is part of the XQuery 3.0 specification. Currently we are using XQuery 1.0 Test Suite.

pavlopoul:
I have checked with the trace function and these entities are indeed added in the final output.

prestoncarman:
While there seems to be some questions on how to print a serialized value, I think this change is a good step in the right direction. Should we merge it as is?

PR #124, opened 2016-07-27
Changes to arithmetic operation semantics

riyafa:


prestoncarman:
The merge had a conflict. It appears github does not recognize that this commit has been merged. Please close.

PR #126, opened 2016-07-27
Implement libjn:accumulate and libjn:intersect

riyafa:


westmann:
I think that there's a bug here. The query 
`libjn:accumulate(({ "a": { "c": "d" } }, { "a": { "c": "d" } }))`
should return
`{"a":[{"c":"d"},{"c":"d"}]}`
as accumulate should not be applied recursively. The current implementation returns 
`{"a":{"c":"d"}}`.
Please add test cases for nested object and non-homogenous sequences containing objects, arrays, and scalar values.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/LibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.vxquery.exceptions.SystemException;
+import org.apache.vxquery.runtime.functions.base.AbstractTaggedValueArgumentScalarEvaluator;
+import org.apache.vxquery.runtime.functions.util.FunctionHelper;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class LibjnAccumulateScalarEvaluator extends AbstractTaggedValueArgumentScalarEvaluator {
```
westmann: Can we share some code with `JnKeysScalarEvaluator` here?

riyafa:
The fact that libjn:accumulate(({ "a": { "c": "d" } }, { "a": { "c": "d" } })) returns {"a":{"c":"d"}} because since the specification for libjn:accumulate was not clear, I thought that it would behave similar to libjn:intersect and added a check to remove duplicate values for the same key when accumulating into an array. So one  { "c": "d" } is removed because it's a duplicate value. 

You can test with an entry like libjn:accumulate(({ "a": { "c": "d" } }, { "a": { "e": "f" } })) which would return {"a":[{"c":"d"},{"e":"f"}]}. I shall add this test case also.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/LibjnAccumulateScalarEvaluator.java
riyafa: I am using a List in JnKeysScalarEvaluator and an array in this class as argument for the removeDuplicates() method. I can change the implementation in JnKeysScalarEvaluator to use an array also if necessary

riyafa: Except for removeDuplicates method I think it would be better to keep both implementations separate because `JnKeysScalarEvaluator` implementation is more efficient. As for the `removeDuplicates` method, shall I move it to FunctionHelper class?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+            tempTvp.getValue(sp1);
+            for (int j = 0; j < sp1.getEntryCount(); ++j) {
+                key = ppool.takeOne(TaggedValuePointable.class);
+                sp1.getEntry(j, tempTvp);
+                addPair(tempTvp);
+            }
+        }
+    }
+
+    protected TaggedValuePointable isDuplicateKeys(TaggedValuePointable key, Set<TaggedValuePointable> pointables) {
```
prestoncarman: Could this method just return a boolean value?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+        tempTvp.getValue(stringKey);
+        op.getValue(stringKey, value);
+        if (tvp == null) {
+            key = ppool.takeOne(TaggedValuePointable.class);
+            key.set(tempTvp);
+            values = new ArrayList<>();
+            values.add(value);
+            tvps.put(key, values);
+        } else {
+            values = tvps.get(tvp);
```
prestoncarman: Could these three lines be:
tvps.get(tvp).add(value);

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+                for (int i = 0; i < sp.getEntryCount(); ++i) {
+                    op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
+                    sp.getEntry(i, tempTvp);
+                    if (tempTvp.getTag() == ValueTag.OBJECT_TAG) {
+                        tempTvp.getValue(op);
+                        addPairs(tempTvp);
+                    }
+                }
+            } else if (arg.getTag() == ValueTag.OBJECT_TAG) {
+                op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
```
prestoncarman: Move the object pointable into the addPairs method. Also, use the pointable pool for its creation or just use a single pointable (final) and reassign its value.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+        TaggedValuePointable tempTvp = ppool.takeOne(TaggedValuePointable.class);
+        try {
+            abvs.reset();
+            ob.reset(abvs);
+            tvps.clear();
+            if (arg.getTag() == ValueTag.SEQUENCE_TAG) {
+                arg.getValue(sp);
+                for (int i = 0; i < sp.getEntryCount(); ++i) {
+                    op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
```
prestoncarman: Move the object pointable into the addPairs method. Also, use the pointable pool for its creation or just use a single pointable (final) and reassign its value.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,170 @@
...
+                }
+                ppool.giveBack(key1);
+            }
+        }
+    }
+
+    protected abstract void addKeyValue(UTF8StringPointable stringKey, TaggedValuePointable taggedValuePointable) throws IOException;
+
+    private void addPair(TaggedValuePointable tempTvp) throws IOException, SystemException {
+        TaggedValuePointable tvp = isDuplicateKeys(tempTvp, tvps.keySet());
```
prestoncarman: As I look at this method I think tvp is just a check if the key exists.
Take a look at this method "getOrDefault(Object key, V defaultValue)".
This could help you simplify this function.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
riyafa: To get the value using a key from the map we would need the exact key in the map, so I am returning the key in the map to get the value using it.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
riyafa: It is not a simple check because we have to use the byte array to see if the keys are equal. And the key that is added the Map could only be used to retrieve the value/s and not any `TaggedValuePointable` which has similar byte array :/

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
riyafa: We have to use a  new ObjectPointable each time to create a new byte array for each key. Cannot use a pointable pool because it also reuses the same pointable.

vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-functions.xml
```diff
@@ -1228,6 +1228,7 @@
         <runtime type="scalar" class="org.apache.vxquery.runtime.functions.json.LibjnValuesScalarEvaluatorFactory"/>
     </function>
 
+<<<<<<< af1ce5a82e6493b2665718667cc3578993a1e0cf
```
prestoncarman: Left over code from a merge???

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,168 @@
...
+    private void addPairs(TaggedValuePointable tvp1) throws IOException, SystemException {
+        op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
+            tvp1.getValue(op);
+            op.getKeys(tvp1);
+            if (tvp1.getTag() == ValueTag.XS_STRING_TAG) {
+                addPair(tvp1);
+            } else if (tvp1.getTag() == ValueTag.SEQUENCE_TAG) {
+                tvp1.getValue(sp1);
+                for (int j = 0; j < sp1.getEntryCount(); ++j) {
+                    key = ppool.takeOne(TaggedValuePointable.class);
```
prestoncarman: key is never used.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    protected abstract void addKeyValue(UTF8StringPointable stringKey, TaggedValuePointable taggedValuePointable)
+            throws IOException;
+
+    private void addPair(TaggedValuePointable tvp1) throws IOException, SystemException {
+        TaggedValuePointable tvp = isDuplicateKeys(tvp1, tvps.keySet());
+        value = ppool.takeOne(TaggedValuePointable.class);
+        tvp1.getValue(stringKey);
+        op.getValue(stringKey, value);
```
prestoncarman, line 131: Maybe pass the ObjectPointable in as a variable. Its unclear what value it holds.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,168 @@
...
+            values.add(value);
+            tvps.put(key, values);
+        } else {
+            tvps.get(tvp).add(value);
+        }
+    }
+
+    private void addPairs(TaggedValuePointable tvp1) throws IOException, SystemException {
+        op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
+            tvp1.getValue(op);
```
prestoncarman: formating.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,168 @@
...
+            values = new ArrayList<>();
+            values.add(value);
+            tvps.put(key, values);
+        } else {
+            tvps.get(tvp).add(value);
+        }
+    }
+
+    private void addPairs(TaggedValuePointable tvp1) throws IOException, SystemException {
+        op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
```
prestoncarman: I don't think this needs to come from a pool. This is not a recursive function.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -0,0 +1,168 @@
...
+                ppool.giveBack(key1);
+            }
+        }
+    }
+
+    protected abstract void addKeyValue(UTF8StringPointable stringKey, TaggedValuePointable taggedValuePointable)
+            throws IOException;
+
+    private void addPair(TaggedValuePointable tvp1) throws IOException, SystemException {
+        TaggedValuePointable tvp = isDuplicateKeys(tvp1, tvps.keySet());
```
prestoncarman, line 128: Could you just call tvps.get(tvp1) and check if the value is null? 
I think the tvps object should tell you if the key exists so you don't have to search it.

riyafa, line 128: tvps is a `Map` and when I call `tvps.put(key, values)`, the key is added to the map by reference. Here tvp1 has a different different reference. Both has to be compared by the content of their byte array and not by their reference. Hence I would need a different method to do that

PR #128, opened 2016-07-28
Implement libjn:descendant-objects and libjn:descendant-pairs

riyafa:


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/LibjnDescendantObjectsScalarEvaluator.java
```diff
@@ -0,0 +1,127 @@
...
+                    sp.getEntry(i, tempTvp);
+                    process(tempTvp);
+                }
+                ppool.giveBack(tempTvp);
+            } else if (tvp.getTag() == ValueTag.OBJECT_TAG) {
+                nestedObject(tvp);
+            } else if (tvp.getTag() == ValueTag.ARRAY_TAG) {
+                insideArray(tvp);
+            } else {
+                XDMConstants.setEmptySequence(tvp);
```
prestoncarman: The else statement does nothing. An empty sequence is returned because the sequence builder never gets a value. Please remove the else statement.

prestoncarman: You can add a test to confirm this works. 
libjn:descendant-objects("no objects")

vxquery-core/src/main/java/org/apache/vxquery/functions/builtin-functions.xml
```diff
@@ -1205,4 +1205,18 @@
         <return type="item()*"/>
         <runtime type="scalar" class="org.apache.vxquery.runtime.functions.json.LibjnFlattenScalarEvaluatorFactory"/>
     </function>
+
+    !-- libjn:descendant-objects($sequence as item()*) as object()* -->
```
prestoncarman: Comment missing starting character.

PR #132, opened 2016-07-31
Implement libjn:project and libjn:remove-keys

riyafa:


vxquery-xtest/src/test/resources/cat/LibrariesInJSONiq.xml
```diff
@@ -40,4 +40,9 @@
       <query name="flatten" date="2016-07-20"/>
       <output-file compare="Text">flatten.txt</output-file>
    </test-case>
+   <test-case name="project" FilePath="Json/Libraries/" Creator="Riyafa Abdul Hameed">
```
prestoncarman, line 43: Add a test case for "string" and () for key values.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/LibjnProjectScalarEvaluator.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.vxquery.datamodel.builders.sequence.SequenceBuilder;
+import org.apache.vxquery.datamodel.values.ValueTag;
+import org.apache.vxquery.exceptions.ErrorCode;
+import org.apache.vxquery.exceptions.SystemException;
+import org.apache.vxquery.runtime.functions.base.AbstractTaggedValueArgumentScalarEvaluator;
+import org.apache.vxquery.runtime.functions.util.FunctionHelper;
+
+import java.io.IOException;
+
+public class LibjnProjectScalarEvaluator extends AbstractTaggedValueArgumentScalarEvaluator {
```
prestoncarman: This method has a lot of similarities to remove. Can you share code between these to methods? Both allow all non objects through with altering their value. The difference lies in what happens when a key is found: one removes the other keeps.

prestoncarman:
Looks good to me. Please squash and prepare for commit.

PR #134, opened 2016-07-31
Implement libjn:values

riyafa:


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/LibjnValuesScalarEvaluator.java
```diff
@@ -0,0 +1,107 @@
...
+            result.set(abvs1);
+        } catch (IOException e) {
+            throw new SystemException(ErrorCode.SYSE0001, e);
+        } finally {
+            ppool.giveBack(tempTvp);
+            ppool.giveBack(sp);
+        }
+    }
+
+    private void addPairs(TaggedValuePointable tempTvp) throws IOException, SystemException {
```
prestoncarman: Is this the same function from SimpleObjectUnionScalarEvaluator? If so, it seems like we should be sharing code here.

prestoncarman: If not, maybe you could pick a better name.

westmann: I think that it's different as the SimpleObjectUnionScalarEvaluator also does duplicate elimination on the keys, but is might be good to call it "getValues(...)"

prestoncarman:
Looks good, squash and prepare for commit.

PR #135, opened 2016-07-31
Updating data handling

pavlopoul:
I thought that it would be more convenient to have all the data types together, so I removed the XML Node Details tab and I moved its contents to data handling and XML data model example tab accordingly. Do you think that this is more readable for the user, or should I keep the old format?


src/site/apt/development_data_handling.apt
```diff
@@ -129,8 +129,46 @@ Developer Data Handling
...
+*-----------------------------------+----------------------+---------------:
+| Comment Node(CNP)                 | {{{https://git-wip-us.apache.org/repos/asf?p=vxquery.git;a=blob;f=vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/nodes/TextOrCommentNodePointable.java}TextOrCommentNodePointable}}      |    1 + length |
+*-----------------------------------+----------------------+---------------:
+| Text Node(TNP)                    | {{{https://git-wip-us.apache.org/repos/asf?p=vxquery.git;a=blob;f=vxquery-core/src/main/java/org/apache/vxquery/datamodel/accessors/nodes/TextOrCommentNodePointable.java}TextOrCommentNodePointable}}      |    1 + length |
+*-----------------------------------+----------------------+---------------:
+
+
+* JSONiq Data Types
+
+  Json-items are used as a data source for JSONiq and must be parsed into Hyracks data. Each json-item defined in JSONiq can be mapped into pointable defined in Apache VXQuery™.
```
prestoncarman: Instead tm, use \x99

westmann:
This approach looks good to me.

westmann:
Could you create a single patch for this?

prestoncarman:
Please include the change to the menu. We need to remove the link to page that is being deleted.

PR #138, opened 2016-08-05
Fix issue in Object key listing and Array unboxing

riyafa:


prestoncarman:
Did you create an issue for this error? Lets link this change to an issue. Otherwise, I think its ready to be squashed and committed.

PR #146, opened 2016-08-16
Adding folder queries_json to benchmark

pavlopoul:


vxquery-benchmark/src/main/resources/noaa-ghcn-daily/queries_json/xml_to_json.xq
```diff
@@ -0,0 +1,35 @@
+(: Licensed to the Apache Software Foundation (ASF) under one
```
prestoncarman: Change file name to q08.xq

vxquery-benchmark/src/main/resources/noaa-ghcn-daily/conf/weather_conf.xml
```diff
@@ -0,0 +1,45 @@
+<!--
```
prestoncarman: This seems to be a specific configuration for your testings. Only include generic examples in the repository.

pavlopoul:
I changed the bash scripts (startcc.sh and startnc.h) because the syntax of the if statement was wrong. When the shortcut of the if statement is used the symbol && should replace the "if then " command. That symbol was missing in the previous implementation.

PR #147, opened 2016-08-18
VXQUERY-227 Refactor ObjectPointable.getKeys()

riyafa:


prestoncarman:
Wow, this was simpler than I expected. Nice work. I have one comment. In several methods, you change the value of an argument passed in and its not related to passing a value back to the caller. I think we should not change a method's argument value unless we are passing a result back to the caller.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnKeysScalarEvaluator.java
```diff
@@ -66,9 +68,10 @@ protected void evaluate(TaggedValuePointable[] args, IPointable result) throws S
                 for (int i = 0; i < size1; i++) {
                     sp1.getEntry(i, temptvp);
                     if (temptvp.getTag() == ValueTag.OBJECT_TAG) {
-                        op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
                         temptvp.getValue(op);
-                        op.getKeys(temptvp);
+                        abvs1 = new ArrayBackedValueStorage();
```
westmann: It would be good to be able to pool the `ArrayBackedValueStorage`s for the evaluator. If we use this evaluator on a big dataset, we'll create and collect the  `ArrayBackedValueStorage`s all the time. I'm not sure if there are other cases in which we already pool those, but I think that it would be very valuable here.

westmann:
Very nice!
To make it even better, I think that it would be good to think about the memory allocation patterns here. I've added a specific comment in one of the relevant places where we allocate  `ArrayBackedValueStorage`s but this also applies to other places.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnKeysScalarEvaluator.java
riyafa: There's a PointablePooleFactory. Am I to create an  ArrayBackedValueStoragesPoolFactory?

prestoncarman:
Another option would be to use one mvs for all keys. Instead of having getKeys reset and start a new key data block, it could just append the key values. The advantage would be you could keep one mvs for all key values. You caller would still have to maintain many key pointables, but only one array backed value store. An easy way to implement this would be to have getKeys take two arguments: IMutableValueStorage mvs and IPointable result. This method would would not reset the mvs and it would give a result pointable for where the appended keys are located.

prestoncarman:
Is the only thing left for this change to start using the new AVBS pool?

riyafa:
Yes it's the only thing left. I have been thinking whether I need to use Singleton pattern to make all classes use the same ArrayBackedValueStoragePool or what design principle is needed to make sure all classes use one instance of the pool rather than creating a new one all the time?

prestoncarman:
For simplicity, I would create a new pool for each evaluator, similar to
the pointable pools. While you could come up with another solution, I think
this works with the framework hyracks provides. The result would be one
pool per function (per partition).

riyafa:
Can this be merged?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/KeysOrMembersScalarEvaluator.java
```diff
@@ -71,6 +71,8 @@ protected void evaluate(TaggedValuePointable[] args, IPointable result) throws S
             }
         } catch (IOException e) {
             throw new SystemException(ErrorCode.SYSE0001, e);
+        }finally {
```
prestoncarman: formating

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/jsonitem/ObjectConstructorScalarEvaluator.java
```diff
@@ -54,6 +52,8 @@ public ObjectConstructorScalarEvaluator(IHyracksTaskContext ctx, IScalarEvaluato
     @Override
     protected void evaluate(TaggedValuePointable[] args, IPointable result) throws SystemException {
         TaggedValuePointable key, value, qmc, tvp;
+        ArrayBackedValueStorage abvs = abvsPool.takeOne();
+        ArrayBackedValueStorage abvs1 = abvsPool.takeOne();
```
prestoncarman: Can we use more descriptive names for the abvs variables? How about abvsResult and abvsItem? Do those make sense?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/JnKeysScalarEvaluator.java
```diff
@@ -39,36 +39,40 @@
...
                 tvp1.getValue(sp1);
                 int size1 = sp1.getEntryCount();
                 int size2;
                 for (int i = 0; i < size1; i++) {
                     sp1.getEntry(i, temptvp);
                     if (temptvp.getTag() == ValueTag.OBJECT_TAG) {
-                        op = (ObjectPointable) ObjectPointable.FACTORY.createPointable();
                         temptvp.getValue(op);
-                        op.getKeys(temptvp);
+                        ArrayBackedValueStorage abvs1 = new ArrayBackedValueStorage();
```
prestoncarman: Is there a more descriptive variable name than abvs1? abvsKeys?

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnProjectScalarEvaluator.java
```diff
@@ -66,6 +63,8 @@ protected void evaluate(TaggedValuePointable[] args, IPointable result) throws S
         }
         TaggedValuePointable tempTvp = ppool.takeOne(TaggedValuePointable.class);
         SequencePointable sp = ppool.takeOne(SequencePointable.class);
+        ArrayBackedValueStorage abvs = abvsPool.takeOne();
+        ArrayBackedValueStorage abvs1 = abvsPool.takeOne();
```
prestoncarman: Similar comment about the abvs variable names.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/json/AbstractLibjnAccumulateScalarEvaluator.java
```diff
@@ -71,6 +71,10 @@ protected void evaluate(TaggedValuePointable[] args, IPointable result) throws S
             throw new SystemException(ErrorCode.FORG0006);
         }
         TaggedValuePointable tempTvp = ppool.takeOne(TaggedValuePointable.class);
+        ArrayBackedValueStorage abvs = abvsPool.takeOne();
+        abvsList.add(abvs);
+        ArrayBackedValueStorage abvs1 = abvsPool.takeOne();
+        abvsList.add(abvs1);
```
prestoncarman: More descriptive abvs variable names.

prestoncarman (changes requested):
I think this change moves us in the direction of having better memory management. Please update the formating and variable naming.

PR #150, opened 2016-08-23
Convert weather data to JSON

pavlopoul:


vxquery-benchmark/src/main/resources/noaa-ghcn-daily/scripts/weather_convert_to_json.py
```diff
@@ -0,0 +1,570 @@
...
+from collections import OrderedDict
+
+# Custom modules.
+from weather_config_ghcnd import *
+from weather_config_mshr import *
+from weather_download_files import *
+
+class WeatherConvertToJSON:
+    
+    STATES = OrderedDict({
```
prestoncarman: Can the STATES and MONTHS variables be shared between convertors?

prestoncarman: It looks like many of the methods here can be shared with the WeatherConvertToXML. It would be nice to have "abstract" class that is shared for each convertor. (Not sure what its called in python.)

vxquery-benchmark/src/main/resources/noaa-ghcn-daily/scripts/weather_convert.py
```diff
@@ -0,0 +1,830 @@
+#!/usr/bin/env python
```
prestoncarman, line 1: Is this a new file or just being renamed?

pavlopoul: It is a new file that contains an abstract converter class. Then, according to which converter we need the methods are implemented in the corresponding converter. The two converters classes are, also contained in this file.

PR #155, opened 2016-09-10
VXQUERY-25: Fixes the file issue with running JUNIT through maven

prestoncarman:
Maven will now be able to run the XQTS JUNIT test without opening
to many files. If XQTS is installed locally then the system will run
these tests during verify and/or integration-tests.

The fix involves only creating one cluster instance, one connection, and
one dataset. These instances are reused for each test.


westmann:
Looks good to me!

PR #158, opened 2016-09-22
Some changes to build the site correctly

pavlopoul:


vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/index/updateIndex/IndexUpdater.java
```diff
@@ -86,12 +86,6 @@ public  IndexUpdater(String indexFolder, IPointable result, UTF8StringPointable
         this.pathsFromFileList = new HashSet<>();
     }
 
-    /**
-     * Perform the initial configuration for index update/ delete processes.
```
prestoncarman, line 90: I think we should keep these comments. They seem to add to the documentation.
Same goes for the next three deleted comments.

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/index/indexCentralizer/IndexCentralizerUtil.java
```diff
@@ -41,7 +41,7 @@
 /**
  * Class for maintaining the centralized index information file.
  * Index centralization procedure.
- * User can specify the collection directory in VXQuery.java, ncConfig.ioDevices = <index_directory>.
```
prestoncarman, line 44: Lets add the HTML less than and greater than symbol. `&lt;`

vxquery-core/src/main/java/org/apache/vxquery/runtime/functions/index/CaseSensitiveAnalyzer.java
```diff
@@ -38,7 +38,7 @@
  * Filters {@link StandardTokenizer} with {@link StandardFilter},
  * and {@link StopFilter}, using a list of
  * English stop words.
- * <a name="version"/>
+ * <a> name="version"</a>
```
prestoncarman: I am not sure why this was not allowed, but this is not a correct fix.
Try this: `<a name="version"></a>`

prestoncarman (changes requested):
A few HTML comment fixes along with a request to keep some of the comments.

PR #160, opened 2016-11-18
depend on com.fasterxml.jackson.core 2.8.4

westmann:
Change the dependency on Jackson from a version that's repackaged in htrace to a regular release.


prestoncarman (approved):
Looks good.

PR #162, opened 2016-11-19
some dependency cleanup

westmann:


prestoncarman (commented):
Can you explain the general reason for these changes? It looks like we have some new dependencies (such as com.google.guava, org.json, log4j) Its a little hard to tell due to the reordering of several dependencies.

westmann:
I used `mvn dependency:analyze` to find out which dependencies we actually had in the code vs which we had declared in the poms. We had some which we didn't declare in the poms but which we used - those only worked because another dependency recursively depended on them. And there were some that we declared but never used. So the ones that are apparently new should be the ones that we used but didn't declare.

prestoncarman:
The change looks good. Did this ever get merged?

prestoncarman:
These changes have been merged. I pull this change and updated it for master. This can be closed.

PR #168, opened 2017-05-12
tillw/hyracks0.3.2

westmann:


prestoncarman:
Hyracks has now pushed its latest release. Can you update this patch to use the last published version?

PR #171, opened 2017-06-04
Prestonc/release prep

prestoncarman:
Started working on site updates based on recent work. This would be a great time to update the site and make a release of our query processor.


src/site/apt/index.apt
```diff
@@ -16,8 +16,9 @@
 Apache VXQuery
 
   Apache VXQuery\x99 will be a standards compliant XML Query processor implemented in Java.
-  The focus is on the evaluation of queries on large amounts of XML data.
-  Specifically the goal is to evaluate queries on large collections of relatively small XML documents.
+  The XQuery processor supports JSONiq extension to XQuery.
```
westmann: I'd call it "the JSONiq extension to XQuery" and I think that we should have a link to jsoniq.org somewhere (maybe here, but not necessarily).

src/site/apt/development_contribution.apt
```diff
@@ -96,11 +100,26 @@ Contributing Code
       At least one other member of the community should review the change.
       If there is any feedback, address this and repeat the posting process.
 
+    * Update your Pull Request.
+
+      Update your change to address any comments from reviewers.
+
+    * Prepare your change for merge.
+
+      Squash your changes into a single commit with a nice commit message.
```
westmann, line 118: It would be really nice if we had a consistent example with command lines to go along with this description - especially for the rebase/squash :)

src/site/apt/development_contribution.apt
```diff
@@ -114,9 +133,30 @@ Contributing Code
...
+    * Double check the change.
+
+      Confirm the change has a single commit and includes a nice commit message.
+
     * Merge the change with ASF master.
 
-      When merging the change, do not rebase.
+      When merging the change, do not ~rebase~.
       Instead do a single merge commit into Apache VXQuery master.
+      Since the Pull Request now has a single commit, yan alternative would be to cherry pick that commit
```
westmann: s/yan/an/

westmann (commented):
Added a few comments. 

On the "architecture diagram", I'd put "XQuery/JSONiq" on top of a single arrow for VXQuery (it's a single language - and it's a pretty useful feature to handle XML and JSON in the same query) while I'd put SQL++ and AQL on different arrows. Also I think that SQL++ should be to the left of AQL.

westmann (approved):
Looks good to me.

PR #176, opened 2017-08-25
POM Dependency Clean Up

prestoncarman:
- updated based on results from: "mvn dependency:analyze"
 - Minor item from GSOC commit.
 - Ordered dependencies to improve maintenance.

This is my attempt to bring the changes in https://github.com/apache/vxquery/pull/162 into master. Something went wrong in the REST project and now the tests fail due to a missing class. Any ideas?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,9 +14,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<project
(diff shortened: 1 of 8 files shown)


pavlopoul:
These changes seem good since there are no dependencies problems.

PR #178, opened 2017-11-04
VXQUERY-241 Refactor fn:contains, fn:ends-with and fn:starts-with

riyafa:


prestoncarman:
Nice, we can actually use the Hyracks string functions. Looks good to merge.

prestoncarman:
The regression tests have a few issues.
https://asterix-jenkins.ics.uci.edu/view/%20VXQuery/job/vxquery-pr/524/

PR #180, opened 2018-07-08
Move Hyracks dependency to 0.3.3.

westmann:
- remove some unused code
- add another check in the HTTP servlet
- nicer QueryRequest.toString


prestoncarman:
Looks good to me. Thanks for moving us to the latest Hyracks release. We have tried several time to do this but it never was finished.

pavlopoul (approved):
Everything seems fine to me.

